PR #13, opened 2021-05-08
add solution

KononenkoKonstantin:


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.debug("login method was called with login: {}", login);
```
divertitoS, line 13: I consider that we should use info method here, but I'll approve this HW and look at the mentor comments

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.debug("completeOrder method was called with userId: {}", userId);
```
divertitoS, line 15: It's the same here)

divertitoS (approved):

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
KononenkoKonstantin: I think we do not need to see this message in info logging level

src/main/java/mate/academy/Main.java
```diff
@@ -6,16 +6,20 @@
...
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
-            return;
+            logger.error("Wrong login or password ", e);
+            throw new RuntimeException("Wrong login or password, e");
```
KseniiaMakarova, line 22: there is no need to log anything if we throw an exception
so let's remove throwing a RuntimeException here

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
KseniiaMakarova, line 23: add empty line to the end of the file

KseniiaMakarova (changes requested):

src/main/resources/log4j2.xml
KononenkoKonstantin: done

KseniiaMakarova (approved):

PR #52, opened 2021-05-12
1ST ATTEMPT

yarrybka:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,28 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method-function completeOrder() was called. Params: userId = {}", userId);
```
LayRipper, line 15: There is a better logger method to use when method invocation

LayRipper (changes requested):

src/main/java/mate/academy/service/OrderServiceImpl.java
yarrybka: debug?

LayRipper: fix it all over the place pls

yarrybka: getAllProductsFromShoppingCart() too?

yarrybka: or i can leave it as info?

LayRipper: except for this one c: We need to know if all the  data was successfully fetched

LayRipper (approved):

Resci (approved):

PR #68, opened 2021-07-08
hw-21 added solution

OlesiaSmitiukh:


Tamara-Paskalova (approved):

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+*.log/*
```
KseniiaMakarova, line 4:
```suggestion
*.log
```

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.debug("Failed authentication");
```
KseniiaMakarova, line 21: should we use debug level here?

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called.");
```
KseniiaMakarova, line 15: this is the information that is helpful for debugging - let's better use a different level and add method parameter to message

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="debug" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
KseniiaMakarova, line 23: add an empty line

KseniiaMakarova (changes requested):

src/main/resources/log4j2.xml
OlesiaSmitiukh: done ✔️

.gitignore
OlesiaSmitiukh: ✔️

src/main/java/mate/academy/service/OrderServiceImpl.java
OlesiaSmitiukh: Edited

src/main/java/mate/academy/Main.java
OlesiaSmitiukh: I didn’t be sure. Is it like "the inability to access a file or service"?

src/main/java/mate/academy/Main.java
```diff
@@ -18,7 +18,7 @@ public static void main(String[] args) {
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            logger.debug("Failed authentication");
+            logger.error("Failed authentication");
```
KseniiaMakarova, line 21: let's also pass the original exception as the second parameter

KseniiaMakarova (approved):

src/main/java/mate/academy/Main.java
KseniiaMakarova: we got an exception in our logic - this means something went wrong, and we should log that some error happened

PR #80, opened 2021-07-12
Fix task description

Resci:


README.md
```diff
@@ -2,6 +2,5 @@
 
 1. Configure logger in this project
 1. Edit `.gitignore` file so that your file with logs is NOT pushed to Github
-1. Replace all `System.out.println()` with corresponding logger methods
```
boroda4436, line 5: Я б переформулював: If you have `System.out.println()` than replace them all with corresponding logger methods

PR #92, opened 2021-09-19
DmytroPetrushenko / jv-logger

DmytroPetrushenko:


src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,21 @@
...
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
+            OrderService orderService = new OrderServiceImpl();
+            orderService.completeOrder(user.getUserId());
         } catch (AuthenticationException e) {
-            e.printStackTrace();
-            return;
+            logger.error("Can't login", e);
```
RomanDubovskyi, line 23:
```suggestion
            logger.warn("Can't login", e);
```
it's at least a warn (error is used when smt goes wrong in the program flow). In most cases invalid password is just a mistype from user

RomanDubovskyi (approved):

PR #94, opened 2021-09-22
Complete program with logging.

osa240:


Sinelife (approved):

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+logs/myLog.log
```
MaliukDaria, line 4:
```suggestion
logs/*
```

I think it's better to exclude the whole logs folder, so if you will add new logger file in the future, you won't need to add changes here

MaliukDaria (changes requested):

.gitignore
osa240: agree)

MaliukDaria (approved):

MaliukDaria (approved):

PR #100, opened 2021-09-24
JV Logger HW

AlexPravda1:
- exceptions without try/catch
- no user names, only ID for security reasons


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,20 @@
...
     public static void main(String[] args) {
+
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
-            user = authenticationService.login("bob", "1234");
+            user = authenticationService.login("bob", "1234ss");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("User login attempt failed", e);
```
RomanDubovskyi, line 22: I would prefer warn level here, since error is mainly used when you have some kind of error in app work. Auth exception is usually just a mistype from user

RomanDubovskyi (approved):

PR #102, opened 2021-09-26
Logger creating

SolnalarmJulia:


Sinelife (changes requested):
1) Didn't see log4j2.xml. Maybe you forgot to add it to commit.
2) Logger also should output info about login method in AuthenticationServiceImpl.

Sinelife (approved):

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+logs/app.log
```
MaliukDaria, line 4: I think it's better to exclude the whole logs folder, so if you will add new logger file in the future, you won't need to add changes here

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #111, opened 2021-09-27
Logger

18Yuliya18:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can't login", e);
```
netiazhevskyi, line 21: I think we should use more informative message in logger.error()

netiazhevskyi (approved):
looks good

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+/logs/*
```
MaliukDaria, line 4:
```suggestion
logs/*
```

src/main/java/mate/academy/Main.java
MaliukDaria: we dont have any incoming parameters here, so we can leave it as it is. (But personally, I used to write something like "User authentication failed. User login or password is incorrect")

MaliukDaria (approved):

PR #118, opened 2021-09-28
logger implemented

Nazar-Hetun:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,18 @@
...
 public class Main {
     public static void main(String[] args) {
+        Logger logger = LogManager.getLogger(Main.class);
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can't authenticate user.", e);
```
andrii-demchenko, line 20: Let's add more info in the message(why we can't?)

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,28 @@
...
-        // NOTE: In production ready code this order identifier should be generated by DB
-        // For test purpose we simplify this and return dummy data
         order.setOrderId(1L);
         return order;
     }
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
+        logger.debug("Data was successfully fetched from DB");
```
andrii-demchenko, line 23: Is debug logging level necessary here. Can't we use info? Also maybe it is better to log it after the data was fetched(after List of products?

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -15,8 +19,8 @@ public User login(String login, String password) throws AuthenticationException
     }
 
     private User findByLogin(String login) {
+        logger.info("User with login: " + login + " successfully found");
```
andrii-demchenko, line 22: If you write that user with login was successfully found, then, maybe, it is better to log it before return statement, when all method logic is complete?

andrii-demchenko (changes requested):

andrii-demchenko (approved):
Looks good

Max-Pochepets (approved):

PR #126, opened 2021-09-28
Develop

Radikhovskiy:


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+logs/app.log
```
kvoytikh, line 4: лучше писать *.log - так будет для всех лог файлов

kvoytikh (approved):
Good job :)

.gitignore
MaliukDaria: I think it's better to exclude the whole logs folder, so if you will add new logger file in the future, you won't need to add changes here

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Incorrect login or password: ", e);
```
MaliukDaria, line 21:
```suggestion
            logger.error("Failed to authenticate user. Incorrect login or password: ", e);
```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
 import java.util.List;
 import mate.academy.model.Order;
 import mate.academy.model.Product;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
```
MaliukDaria, line 11: why AuthenticationServiceImpl?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #137, opened 2021-10-01
Hw 15 solution

7777vova7777:


src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
RomanDubovskyi, line 23: empty line is missing

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can't authenticate user.", e);
```
RomanDubovskyi, line 21: let's use warn here, since error usually indicates about code failures, in this case usually it's just a mistype

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #180, opened 2021-12-02
added logger (log4j)

MaksymMazlov:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can`t found login", e);
```
RomanDubovskyi, line 21: I would use warn here, since error usually indicates that program didn't work as intended, here we probably have just user mistake

RomanDubovskyi (approved):

PR #185, opened 2021-12-14
solving hw-15-solution jv-logger

oleh-mazurkov:


src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
RomanDubovskyi, line 23: missing an empty line at the end

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
-            user = authenticationService.login("bob", "1234");
+            user = authenticationService.login("bob", "1234222");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can't login", e);
```
RomanDubovskyi, line 21: let's use warn here, since if use mistypes his password it doesn't mean that program works incorrectly

RomanDubovskyi (changes requested):

src/main/java/mate/academy/Main.java
oleh-mazurkov: no feeling such subtleties)) need experience...

RomanDubovskyi: all good, that's not obvious thing if you haven't met it )

RomanDubovskyi (approved):

PR #189, opened 2022-01-17
Hw 15 solution

Viktor-Stadnik:


logs/application.log
```diff
@@ -0,0 +1,3 @@
+2022-01-17 16:22:47,477 INFO  m.a.s.AuthenticationServiceImpl.java:14 login method was called. Login: bob
```
MaliukDaria, line 1: task: Edit .gitignore file so that your file with logs is NOT pushed to Github

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+                <AppenderRef ref="LogToFile"/>
+                <AppenderRef ref="LogToConsole"/>
+        </Logger>
+
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
MaliukDaria, line 23: don't forget about empty lines in the end of file

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #195, opened 2022-02-05
Jv logger

araslav:


src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="INFO">
+            <AppenderRef ref="LogToFile" />
+            <AppenderRef ref="LogToConsole" />
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile" />
+            <AppenderRef ref="LogToConsole" />
+        </Root>
+    </Loggers>
+</Configuration>
```
RomanDubovskyi, line 23: missing an empty line at the end of the file

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("User not login", e);
```
RomanDubovskyi, line 21: I would rather use warn here since the program works fine, just a user made a mistype in the login or pwd

RomanDubovskyi (changes requested):

Resci (approved):

PR #214, opened 2022-02-23
Completed Logger tasks

DenTima:


src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,21 @@
...
 public class Main {
+    public static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
+            OrderService orderService = new OrderServiceImpl();
+            orderService.completeOrder(user.getUserId());
```
NazariiYatsiuk, line 21: There is no logic to move this strings into try/catch block, because AuthenticationException can be called only by authenticationService.login() method.

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,17 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager
+            .getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
         //TODO: add corresponding log message about method login was called
```
NazariiYatsiuk, line 14:
```suggestion
```
Lets remove this and lower comments

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -25,6 +30,7 @@ public Order completeOrder(Long userId) {
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
         // TODO: add log message about successful fetched data from DB
+        logger.info("data was fetched from DB");
```
NazariiYatsiuk, line 33: Logger message will be more informative if you mention ID of user, with which this method was succesfully performed

NazariiYatsiuk (changes requested):

src/main/java/mate/academy/Main.java
MaliukDaria: Agree, lets remove it from here

src/main/java/mate/academy/service/OrderServiceImpl.java
MaliukDaria: agree

KostyaNikiforov (approved):

PR #227, opened 2022-03-19
Logger was successfully implemented, some classes were edited

liubcheck:


src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,39 @@
...
+                        break;
+                    case 1:
+                        user = authenticationService.login("jake", "1234");
+                        break;
+                    case 2:
+                        user = authenticationService.login("bob", "qwerty");
+                        break;
+                    default:
+                        break;
+                }
```
MaliukDaria, line 36: why do we need this loop and switch-case construction?

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,39 @@
...
-            user = authenticationService.login("bob", "1234");
-        } catch (AuthenticationException e) {
-            e.printStackTrace();
-            return;
-        }
+        logger.info("AuthenticationService object was created");
+        User user = null;
         OrderService orderService = new OrderServiceImpl();
-        orderService.completeOrder(user.getUserId());
+        logger.info("OrderService object was created");
```
MaliukDaria, line 21: I dont think we ned to log object creation

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="Console"/>
+            <AppenderRef ref="File"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="Console"/>
+            <AppenderRef ref="File"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
MaliukDaria, line 23: dont forget about empty lines in the end of file

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #241, opened 2022-05-03
jv_homework_logger

etezaadi:
- configured logger;
- replaced e.printStackTrace() with logger method;
- ignored logs file.


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
 import java.util.List;
 import mate.academy.model.Order;
 import mate.academy.model.Product;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger();
```
MaliukDaria, line 11:
```suggestion
    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
```

MaliukDaria (changes requested):

src/main/java/mate/academy/service/OrderServiceImpl.java
etezaadi: @MaliukDaria already fixed :)

MaliukDaria (approved):
👍🏻

PR #275, opened 2022-05-28
added solution for jv-logger task

dmnichz:


src/main/java/mate/academy/Main.java
```diff
@@ -6,16 +6,19 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class Main {
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
-        User user;
+        User user = null;
```
MaliukDaria, line 17: why do we need this initialization?

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
MaliukDaria, line 23: dont forget about empty line in the end of file

MaliukDaria (changes requested):

src/main/java/mate/academy/Main.java
dmnichz: image
there is a compilation error without initialization

PR #285, opened 2022-06-14
implement task

gdpbalt:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.debug("Method completeOrder was called. Params: userId={}", userId);
```
Sm0kk, line 15: I think here we must used .info. Because info used when your program is working as expected, but debug used when we find the reason in case your program is not working as expected.

Sm0kk (changes requested):

Sm0kk (approved):

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Authentication error", e);
```
KostyaNikiforov, line 21: Let's add information about the user who got an error

src/main/java/mate/academy/model/Product.java
```diff
@@ -29,4 +30,12 @@ public BigDecimal getPrice() {
     public void setPrice(BigDecimal price) {
         this.price = price;
     }
+
+    @Override
+    public String toString() {
+        return new StringJoiner(", ", Product.class.getSimpleName() + "[", "]")
+                .add("title='" + title + "'")
+                .add("price=" + price)
+                .toString();
```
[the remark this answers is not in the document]
gdpbalt: I agree with you
I didn't think about that case

KostyaNikiforov (approved):

PR #324, opened 2022-08-10
Added dependency for log4j logger and resolved all tasks

vitaliy1210:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can't log in user with login: {}; and password: {}.", "bob","1234");
```
MaliukDaria, line 21: I think we should not log info about password, it is sensitive data

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,18 @@
...
+    </Appenders>
+    <Loggers>
+        <Logger name="mate.academy" level="debug">
+            <AppenderRef ref="LogToFile"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
MaliukDaria, line 18: dont forget about empty line in the end of file

MaliukDaria (changes requested):

src/main/java/mate/academy/Main.java
vitaliy1210: Yes, I thought that it will useful for client registration, but  I did't count that exception will only by the invalid login

zachar-chorny (approved):

PR #347, opened 2022-09-04
Init commit

max-bolotin:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called");
```
vadimkik2, line 15: Also you may add more information in your logger, like:
logger.info("Method completeOrder was called. Params: userId = {}", userId)

I think it's not private info.

vadimkik2 (approved):

Sofasmile (approved):

PR #365, opened 2022-09-05
add solution

AndrewHudzenko:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,29 @@
...
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
+        logger.info("getAllProductsFromShoppingCart method was called. Params: userId = {}",
+                userId);
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("successful fetched data from DB. Params: products = {}", products);
```
valentyn-slutskyi, line 29:
```suggestion
        logger.info("successful fetched data from DB. Params: userId = {}", userId);
```
I do not think logging DB is a good idea. What if DB will contain some secret data?
I think it will be better to log just income params. This way we can know who have access to the DB and will find the traitor)

valentyn-slutskyi (changes requested):

src/main/java/mate/academy/service/OrderServiceImpl.java
AndrewHudzenko: I think it depends on the case, here we have an order, no secret data, I'll leave it at that

valentyn-slutskyi (approved):

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+logs/*
```
Sofasmile, line 4: add empty line at the end of the file

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
Sofasmile, line 23: add empty line at the end of the file

Sofasmile (changes requested):

Sofasmile (approved):

PR #366, opened 2022-09-05
Was added Logger to the project

valentyn-slutskyi:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -25,6 +30,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
         // TODO: add log message about successful fetched data from DB
+        logger.info("Data was successful fetched from DB. Params: userId = {}", userId);
```
AndrewHudzenko, line 33: i guess here u shoud call products , and if u will call products u need to override method toString in product class/ and maybe u need to delete all comments ?

AndrewHudzenko (changes requested):

src/main/java/mate/academy/service/OrderServiceImpl.java
valentyn-slutskyi: I do not think logging DB is a good idea. What if DB will contain 1K lines? 10K? 100K and even more?

Noct2000 (changes requested):
add `log4j2.xml` to commits

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
Noct2000, line 23: add new line to the end of the file

Noct2000 (approved):

PR #370, opened 2022-09-05
Logger is completed

Kupriiuk:


dhorunka (approved):

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+/logs/app.log
```
Noct2000, line 4: I think it would be better 
```suggestion
logs/*
```

Noct2000 (approved):

PR #434, opened 2022-11-08
implemented log4j2 logger

muzykaVP:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error(e);
```
YuliaTsariv, line 21: Perhaps it would be better to use a more informative error?

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
+*.log
```
YuliaTsariv, line 3: logs/*

YuliaTsariv (approved):

pom.xml
```diff
@@ -17,14 +17,28 @@
...
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.antlr</groupId>
+            <artifactId>antlr4-runtime</artifactId>
+            <version>4.8-1</version>
+        </dependency>
```
hronograf, line 41: do we need all these libraries?
looks like this should be enough:
```
<dependency>
      <groupId>org.apache.logging.log4j</groupId>
      <artifactId>log4j-core</artifactId>
      <version>2.19.0</version>
</dependency>
```

hronograf (changes requested):

pom.xml
muzykaVP: i found it on log4j2 site and copy, i know that in lesson we only use .core one, but i wasn`t sure, so i added those also

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Data was successfully fetched from DB. Params: userID = {}", userId);
```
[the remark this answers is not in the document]
hronograf: image

hronograf (changes requested):

hronograf (approved):
approved
and @muzykaVP we could also override `toString` for `Product` and just log `products`

PR #436, opened 2022-11-08
Logger was added

AlinaOryshak:


anton-pushkar (approved):

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -16,7 +20,6 @@ public User login(String login, String password) throws AuthenticationException
 
     private User findByLogin(String login) {
         User user = new User(login, "1234");
-        // this user identifier should be set by DB. We will use dummy data for this example
         user.setUserId(2L);
```
liliia-ponomarenko, line 23: Let’s  add a log that the method findByLogin(String login) was called

liliia-ponomarenko (changes requested):
Great!  let’s  add one more log

liliia-ponomarenko (approved):
LGTM  🔥

PR #464, opened 2022-12-16
first commit on task jv-logger

romanovosad87:


zachar-chorny (changes requested):
Where your log4j2.xml file?)

zachar-chorny (approved):

PR #474, opened 2022-12-27
Implemented logger

SofiiaKazanivska:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
-            user = authenticationService.login("bob", "1234");
+            user = authenticationService.login("bob", "12345");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can not authenticated", e);
```
Rommelua, line 21:
```suggestion
            logger.error("Can not authenticate", e);
```

src/main/resources/resources/log4j2.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="info">
```
Rommelua, line 2: fix alignment in this file

Rommelua (changes requested):
Edit .gitignore file so that your file with logs is NOT pushed to Github

Rommelua (approved):

PR #500, opened 2023-01-18
Logger task solved 1 commit

ar4ibaldd:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
 import mate.academy.service.OrderService;
 import mate.academy.service.OrderServiceImpl;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class Main {
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
```
Dastya, line 16: You can do this field constant

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can`t login", e);
             return;
         }
         OrderService orderService = new OrderServiceImpl();
```
Dastya, line 24: And this field too

Dastya (changes requested):

src/main/java/mate/academy/Main.java
ar4ibaldd: this field already constant

Dastya (approved):

Nick97-git (approved):

PR #520, opened 2023-01-31
Сonfigured logger. Edited .gitignore. TODO done.

andrey156:


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+logs/*
```
Rommelua, line 4: image

Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

.gitignore
andrey156: I thought that there is no need to leave)

Rommelua (approved):

PR #529, opened 2023-02-10
jv-logger-solution

nkharyshyn:
Configured logger, ignored file with logs, completed all TODO statements, replaced e.printStackTrace


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Username or password are incorrect");
```
Ivan95kos, line 21: We should see the stacktrace of the error
```suggestion
            logger.error("Username or password are incorrect", e);
```

Ivan95kos (changes requested):

aleksandr-bevza (approved):

PR #535, opened 2023-02-21
Configure logger

katerynapoli:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("Data was successfully fetched. Params: userId={}", userId);
```
olekskov, line 27: I believe here is better to add details regarding fetched data, not the user (or not only the user).
When something went wrong you will open log and see something like:
   CompleteOrder method was called. Params: userId=2
   Data was successfully fetched. Params: userId=2
And it won't be an easy task to understand what went wrong and why customer is angry

olekskov (changes requested):
Nice, but requires additional details logging

okuzan (approved):
well done

PR #536, opened 2023-02-21
jv-logger solution was created.

mrGvvD:
Edited pom.xml
created log4j2.xml
Added logging of exception and methods.


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.warn("Impossible to authenticate User");
```
Ivan95kos, line 21: We need to see the stack trace of the error
```suggestion
            logger.warn("Impossible to authenticate User", e);
```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 import mate.academy.model.Product;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
```
Ivan95kos, line 11: You need to add a log here

Ivan95kos (changes requested):

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("All products from shopping car has been got for userId: {}", userId);
```
olekskov, line 27: We may also want to add products logging in order to be able to obtain maximum info during log checks.
We check logs in cases when we would like to have maximum information possible (usually when something is already wrong and we need to fix it)

olekskov (approved):
Good job, please check comment for further consideration

PR #543, opened 2023-03-04
Added solution for  jv-logger

serhii-shyian:


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+*.log
```
kozhukhovsky, line 4:
```suggestion
*.log
logs/
```

logs/AuthenticationService.log
```diff
@@ -0,0 +1,7 @@
+2023-03-04 10:18:39,004 INFO mate.academy.service.AuthenticationServiceImpl:14 login method was called!. Params: login=bob
```
kozhukhovsky, line 1: folder with logs shouldn't be in repo, remove pls

kozhukhovsky (approved):

PR #546, opened 2023-03-11
I configure logger in this project

NazarZva:


mishok321 (approved):

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToConsole"/>
+            <AppenderRef ref="LogToFile"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToConsole"/>
+            <AppenderRef ref="LogToFile"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
nacenik, line 23: add an empty line to the end of the file. [See common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/logger/logger#common-mistakes)

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Problem with user authentication: " + e);
```
nacenik, line 21: [You should put an exception as a second value into `error` method.](https://logging.apache.org/log4j/2.x/manual/logbuilder.html#log-builder) Not an additional part of the message

nacenik (changes requested):
Good solution! Let’s improve it 😉

PR #548, opened 2023-03-13
logger configured

oleksandr-petrunovskyi:


Rommelua (changes requested):
Fix class Main as well
image

PR #557, opened 2023-03-19
Hw-logger

KosMarch:


kromch (approved):
normal
but you would thinking about my remark

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called. UserId: {}", userId);
```
KseniiaMakarova, line 15: info is ok here as well, no need to change to debug

src/main/java/resources/log4j2.xml
KseniiaMakarova, line 1: file placement is not ok
should be src/main/resources

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #559, opened 2023-03-19
added logger

PlagueEbola:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,28 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.debug("Method completeOrder was called. UserId = {}", userId);
```
ArtemGrunin, line 15: in my opinion it's better to use logger.info here because you still have no problem)

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,28 @@
...
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("Data was successfully retrieved from the DB."
+                + " User ID = {}, Quantity of products = {}", userId, products.size());
```
ArtemGrunin, line 28: and here also) 
we use logger.debug when something has already gone wrong. but only for information it is better to use info

ArtemGrunin (approved):

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,28 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("Data was successfully retrieved from the DB."
```
Andriy-dz, line 27: here I would use info because we are just saying that the data has been loaded. And in the previous cases, I agree with you that you need to use debug because we are saying that some method was called

Andriy-dz (changes requested):
Good Job!

PR #560, opened 2023-03-19
Add logger

VitaliiPotseluiko:


Nikitos787 (approved):

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Login method was called. Params: login={}", login);
```
maxline, line 13: Please use debug level of logging for  the  such message as  which method was called.

maxline (changes requested):
Good job! let's improve the solution

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("User was successful fetched from DB by userId={}", userId);
```
Andriy-dz, line 27: The team and I consulted and decided that it is better to use info here, because it is a normal valid method behavior

Andriy-dz (approved):

PR #562, opened 2023-03-20
solved task

KyslyiViacheslav:
Configured logger
System.out.println() replaced by corresponding logger methods Completed all TODO statements
Replaced all e.printStackTrace(); with corresponding logger methods


AlexandLes (approved):
well done!

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("Successful fetched data from DB. Parameter userId: {}", userId);
```
KseniiaMakarova, line 31: debug level is used when indicating that some method was called or in other debugging issues
info level is used to track business logic. here it is important for business logic that products were fetched from DB
```suggestion
        logger.info("Successful fetched data from DB. Parameter userId: {}", userId);
```

KseniiaMakarova (changes requested):

PR #568, opened 2023-03-20
added homework

sanya-1xx8:


src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
ArtemGrunin, line 23: you need +1 line in the end of this file

pom.xml
```diff
@@ -25,6 +25,16 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+            <version>2.20.0</version>
+        </dependency>
```
KseniiaMakarova, line 32: do we need 2 dependencies or only logj4-core?

KseniiaMakarova (approved):

pom.xml
sanya-1xx8: Yes, we can use only logj4-core) My bad

PR #570, opened 2023-03-20
test

NickolayVoloshchuk:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Successful fetched data from DB. User ID: {}.", userId);
```
Nimakel, line 27: In my opinion. Redundant Info about user id

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} %-5level %c:%L - %msg%n"/>
+        </Console>
+        <File name="LogToFile" fileName="logs/app.log">
+            <PatternLayout>
+                <Pattern>%d %p %c:%L %m%n</Pattern>
+            </PatternLayout>
+        </File>
+    </Appenders>
+    <Loggers>
+        <Logger name="mate.academy" level="debug" additivity="false">
```
Nimakel, line 14: In my opinion we should use info level, not debug
```suggestion
        <Logger name="mate.academy" level="info" additivity="false">
```

Nimakel (approved):
Overall good job, No major problems to fix. Let's see what mentors will say

src/main/resources/log4j2.xml
NickolayVoloshchuk: I think it's an optional value ;)
I experimented a bit with these values. And I left it as it was since we are not forced anywhere in this project to use one or another level of logging)

src/main/java/mate/academy/service/OrderServiceImpl.java
maxline, line 27: In my opinion, this is more of a debug message, since we are checking what was successful and what was not

maxline (approved):
Well done! Approve with small comment, please take a look for the future implementation

PR #571, opened 2023-03-20
Maksym Yashyn Homework

Nimakel:


NickolayVoloshchuk (approved):

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("User authentication service have problem: ", e);
```
maxline, line 21: In English we should use `has` instead `have` in this case. But it is better to use such messages as
"Login failed. Authentication failed. Can't log in. Login or password was incorrect.`

maxline (approved):
Well done! Approved with a small comment, please take a look for the future implementation.

PR #577, opened 2023-03-21
jv-logger solution

potrap:


HelenMatus (approved):
Good solution)

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -2,29 +2,32 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Data successful fetched from DB. Data is: {}", () -> products.stream()
```
maxline, line 28: this is more debug message then info,  since we are checking what was successful and what was not

maxline (changes requested):
Good job. Left just one small comment, take a look please

PR #578, opened 2023-03-21
Implemented solution for hw.

RandomEastEcho:


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.debug("Method login was called. Params: login={}", login);
```
ArtemGrunin, line 13: in my opinion it's better to use logger.info here because you still have no problem)

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.debug("completeOrder method was called. Params: userId = {}", userId);
```
ArtemGrunin, line 15: and here also)
we use logger.debug when something has already gone wrong. but only for information it is better to use info

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -25,6 +29,8 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
         // TODO: add log message about successful fetched data from DB
+        logger.debug("getAllProductsFromShoppingCart finished work successful."
+                + " fetched data from DB. Params: userId = {}", userId);
```
ArtemGrunin, line 33: same moment)) but if you are debugging this code now, let it be))

ArtemGrunin (approved):

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -25,6 +29,8 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
         // TODO: add log message about successful fetched data from DB
+        logger.debug("getAllProductsFromShoppingCart finished work successful."
```
KseniiaMakarova, line 32: debug level is used when indicating that some method was called or in other debugging issues
info level is used to track business logic. here it is important for business logic that products were fetched from DB
see TODO
```suggestion
        logger.info("Successfully fetched data from DB"
```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -25,6 +29,8 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
         // TODO: add log message about successful fetched data from DB
```
KseniiaMakarova, line 31: remove TODO

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #589, opened 2023-03-21
implemented logger

SanykVolodymyr:


pechivo (approved):

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("Data was fetched from database. Params: user Id: {}", userId);
```
Andriy-dz, line 31: in this case, it is better to use the info level, because you are simply informing about the valid operation of the method

Andriy-dz (approved):
Good Job!

PR #590, opened 2023-03-21
added logger to project

outref:


PlagueEbola (approved):
I approve,  but please add fix before mentor review

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,18 @@
...
 
 public class Main {
     public static void main(String[] args) {
+        Logger logger = LogManager.getLogger(Main.class);
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
-        User user;
+        User user = new User("Bob", "1234");
         try {
-            user = authenticationService.login("bob", "1234");
+            user = authenticationService.login(user.getLogin(), user.getPassword());
```
denys-mg, line 18: I think it's redundant to create user object here, as you will receive it from our fake DB (method findByLogin). As a result, you only need to pass some params

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("Data successful fetched from DB");
```
denys-mg: I think here logging level  'info' will be more appropriate

Andriy-dz (approved):
Good Job!

denys-mg (changes requested):
Good job, left you my remarks

PR #597, opened 2023-04-03
Logger task solution.

MelnychukMyhailo:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Data from DB was successful fetched.");
```
olekskov, line 31: Please add information of what was fetched. Logs are required to find root cause of some unpredicted problem and not always we may fully reproduce whole steps list so can count only on what is already logged

olekskov (changes requested):

okuzan (approved):

PR #599, opened 2023-04-03
solved task - configured logger

yaaariiik:


pom.xml
```diff
@@ -25,7 +25,13 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.19.0</version>
+        </dependency>
     </dependencies>
+
```
uliana-andrushchuk, line 34:
```suggestion
```
remove empty line

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger LOGGER = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        LOGGER.info("Method completeOrder was called. Params: userId - {}", userId);
```
uliana-andrushchuk, line 15: `debug` level is used when indicating that some method was called or in other debugging issues.
`info` level is used to track business logic. So, I'd recommend using debug here

uliana-andrushchuk (approved):

PR #602, opened 2023-04-05
Artur

arthursilver111:


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("login method was called. Params: login = {}",login);
```
uliana-andrushchuk, line 13: `info` level is used to track business logic. 
`debug` level is used when indicating that some method was called or in other debugging issues.

So, I'd recommend using debug here

uliana-andrushchuk (approved):

PR #604, opened 2023-04-07
added proper solution for this task

polikarpov13:


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,18 @@
 
 import mate.academy.exception.AuthenticationException;
 import mate.academy.model.User;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger authenticationServiceImplLogger
```
olekskov, line 9: there is no real sense to give long name for logger, the way you dit it in Main class is fine

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -25,6 +32,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
         // TODO: add log message about successful fetched data from DB
+        orderServiceImplLogger.info("Data was successfully fetched from DB");
```
olekskov, line 35: Please add information regarding fetched data. We most commonly check logs when there is a problem reported and we require as much info as possible to find, reproduce and fix bug asap

olekskov (changes requested):

olekskov (approved):
Good job

PR #612, opened 2023-04-23
Configured logger level in task

yaroslav-ometiukh:
-added dependencies to pom.xml
-added log4j2.xml config file


Ivan95kos (approved):
Great job

PR #613, opened 2023-04-25
Hw jv logger ws2019

iamchgi:


olekskov (approved):
Good job. Approved.
I'd log full details about fetched data, not only quantity to have more information for the case of some problems reported

PR #618, opened 2023-05-08
jv-logger-solution

Nikostrik:


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+log4j2.xml
```
Rommelua, line 4: log4j2.xml we need)) commit it.
```suggestion
logs/*
```

Rommelua (changes requested):

PR #622, opened 2023-05-12
completed task

sZlishchev:


OlehPryshchepa (changes requested):
great job, just fix it so that the logs don't get to the github

OlehPryshchepa (approved):
Great job, keep pushing:)

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Login method was called. Params: login={}", login);
```
maxline, line 13: Let's make this log message with level `debug`. Info about called methods usually needed only for debug purposes. If in real application we put info about all methods called to the info level it will increase size of the log files dramatically.

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called. Params: userId={}", userId);
```
maxline, line 15: Let's make this log message with level `debug`. Info about called methods usually needed only for debug purposes. If in real application we put info about all methods called to the info level it will increase size of the log files dramatically.

maxline (changes requested):
Good job! Let's improve the solution

src/main/java/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToConsole"/>
+            <AppenderRef ref="LogToFile"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToConsole"/>
+            <AppenderRef ref="LogToFile"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
sarakhmen, line 23: do not also forget about empty line at the end of the file here :)

sarakhmen (approved):
Good job ;)

PR #623, opened 2023-05-12
solved task

OlehPryshchepa:


sZlishchev (approved):
Great job! Keep pushing))

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger LOGGER = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        LOGGER.info("login method was called. With params login: {}", login);
```
maxline, line 13: Let's make this logger message with level `debug`. Info about called methods usually needed only for debug purposes. If in real application we put info about all methods called to the info level it will increase size of the log files dramatically.

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger LOGGER = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        LOGGER.info("completeOrder method was called. With params userId: {}", userId);
```
maxline, line 15: Let's make this logger message with level `debug`. Info about called methods usually needed only for debug purposes. If in real application we put info about all methods called to the info level it will increase size of the log files dramatically.

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +24,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        LOGGER.info("Successful fetched data from DB. With params userId: {}", userId);
+        LOGGER.debug("Successful fetched data from DB. With params userId: {}", userId);
```
sarakhmen, line 27: info level is used to track business logic. here it is important for business logic that products were fetched from DB

sarakhmen (changes requested):
let's slightly improve your solution one more time :)

PR #630, opened 2023-05-24
HW-jv-logger-24-Of-May

ViktoriiaSvidovska:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,29 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Data from the database was successfully fetched. UserId: ", userId);
```
olekskov, line 29: 1. Use '{}' placeholder for userId in logger message: logger.info('Data from the database was successfully fetched. UserId: {}', userId);
2. Please recheck other logging calls
3. Please also add here information regarding data fetched from DB for better debugging experience

olekskov (changes requested):

.gitignore
```diff
@@ -1,3 +1,5 @@
 .idea/*
 *.iml
 target/*
+src/main/java/java/resources/*
```
overpathz, line 4: In general, it is considered a good practice to include resources to gitignore file. But, in the task you have to create log4j2.xml/properties in this folder and we need to review it. Since you have included this file, I cannot do this.

src/main/java/mate/academy/app/Main.java
```diff
@@ -1,20 +1,24 @@
...
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
-        } catch (AuthenticationException e) {
-            e.printStackTrace();
+        } catch (AuthenticationException exception) {
+            logger.info("Login failed", exception);
```
overpathz, line 21: I consider here is should be error log level

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,16 @@
...
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger
+            = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Method login was called. Login: {}", login);
```
overpathz, line 14: Debug level is used when indicating that some method was called or in other debugging issues. Info level is used to track business logic. Here, we just log that something happen to debug, so..

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,29 @@
...
+            = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
         List<Product> products = getAllProductsFromShoppingCart(userId);
         Order order = new Order(products, userId);
-        // NOTE: In production ready code this order identifier should be generated by DB
-        // For test purpose we simplify this and return dummy data
+        logger.info("Method completeOrder was called. UserId: {}. Products: {}", userId, products);
```
overpathz, line 18: Here is the same with debug level and move it on the start of method. Because, if some error will happen we wont even see the log message.

overpathz (changes requested):
In general everything is okay. But you need a little changes to approve.

okuzan (changes requested):
You didn't commit `log4j2.xml`. Add it to git `git add -f path/to/ignored/file`

src/main/java/mate/academy/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Rommelua, line 1: This file has wrong location. It is not java
image

Rommelua (changes requested):

Ivan95kos (approved):
For the future, do not write changes in the master branch, create a separate branch for this

PR #634, opened 2023-05-30
hw-jv-logger

VolodymyrDavidov:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
 import mate.academy.service.AuthenticationServiceImpl;
 import mate.academy.service.OrderService;
 import mate.academy.service.OrderServiceImpl;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class Main {
+    private static final Logger LOGGER = LogManager.getLogger(Main.class);
```
okuzan, line 13: the naming convention for constants (CONSTANT_CASE) is not typically applied to logger instances because they are not treated as true constants in the traditional sense. Instead, they are considered variables that reference logging instances

okuzan (approved):
Good job

PR #638, opened 2023-05-31
Configured logger in this project.

V-Solianyk:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("CompleteOrder method was called. Parameter: userId : {}", userId);
```
sarakhmen, line 15: Let's make this logger message with level debug. Info about called methods is usually needed only for debugging purposes. If in a real application, we put info about all methods called to the info level it will increase the size of the log files dramatically.

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("login method was called. Params: login : {}", login);
```
sarakhmen, line 13: the same issue here

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,24 @@
...
+            <Logger name = "mate.academy" level = "info" additivity = "false">
+            <AppenderRef ref="Console"/>
+            <AppenderRef ref="File"/>
+       </Logger>
+            <Root level = "error">
+            <AppenderRef ref="Console"/>
+            <AppenderRef ref="File"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
sarakhmen, line 24: let's also add an empty line to the end of this file

sarakhmen (changes requested):
Let's slightly improve your solution :)

sarakhmen (approved):
Good job :)

PR #647, opened 2023-06-06
add logger solution

AnastasiiaVariuchuk:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -2,29 +2,32 @@
...
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Successful fetched data - {} from DB: ",
+                products.stream().map(p -> p.getTitle() + " " + p.getPrice())
+                        .collect(Collectors.toList()));
```
ArinaPolna, line 30: Cood job) Just a one comment: Firstly it's better to use debug level here(because we  are using this level when indicating that some method was called) Secondly it's better to show input parameters(userId) in our logger message)

ArinaPolna (approved):

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Incorrectly entered login: ", e);
```
rovein, line 21:
```suggestion
            logger.error("Incorrectly entered login", e);
```

If there are no additional details, leave message like this.

src/main/java/mate/academy/service/OrderServiceImpl.java
rovein: Good, it is optional to log the products, it would be enough to log just `userId`.

rovein (approved):
Great job! Left minor comments.

PR #651, opened 2023-06-06
create logger for authentication and order services

ArinaPolna:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,28 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("getAllProductsFromShoppingCart method successfully "
```
AnastasiiaVariuchuk, line 27: I recommend using this output, it will be more informativelogger.info("Successful fetched data - {} from DB: ",
                products.stream().map(p -> p.getTitle() + " " + p.getPrice())
                        .collect(Collectors.toList()));

AnastasiiaVariuchuk (commented):
Good job, just minor comment))

src/main/java/mate/academy/service/OrderServiceImpl.java
ArinaPolna: imho it wiil be better to show input parameters in our logger message(so i will leave my solution)

AnastasiiaVariuchuk (approved):

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,28 @@
...
+                     + "fetched data from DB. UserId = {}", userId);
```
rovein, line 28:
```suggestion
    logger.info("Successfully fetched data from DB. Params: userId={}", userId);
```

Let's use `info` log level here, cause we are informing about the business logic flow.

rovein (changes requested):
Great job! Let's improve one point

rovein (approved):
Great job!

PR #664, opened 2023-06-07
Added task solution

pkobevko:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
 import mate.academy.service.AuthenticationServiceImpl;
 import mate.academy.service.OrderService;
 import mate.academy.service.OrderServiceImpl;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class Main {
+    private static final Logger LOGGER = LogManager.getLogger(Main.class);
```
m-pnsk, line 13: Logger is not a typical constant. I think it should be written in lower case. But that's up to you.

m-pnsk (approved):
Good job

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,24 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"></AppenderRef>
+            <AppenderRef ref="LogToConsole"></AppenderRef>
+        </Root>
+    </Loggers>
+</Configuration>
```
sarakhmen, line 24: Let's add an empty line to the end of this file

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #672, opened 2023-06-11
Work ith logger.

Hlib13:


src/main/java/mate/academy/exception/AuthenticationException.java
```diff
@@ -1,8 +1,13 @@
 package mate.academy.exception;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
 public class AuthenticationException extends Exception {
+    private static final Logger logger = LogManager.getLogger(AuthenticationException.class);
```
olekskov, line 7: Please remove logger from Exception class. We log error during handling of exceptions, not during creation

src/main/java/mate/academy/model/Order.java
```diff
@@ -1,8 +1,11 @@
 package mate.academy.model;
 
 import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class Order {
+    private static final Logger logger = LogManager.getLogger(Order.class);
```
olekskov, line 8: Same for all models, please remove, logging required during some business logic flow and model should not help service to perform logging.
Let's imagine we have 100 classes that uses user and we wish to add logging for all of them. Our User class will have 100 different methods for proper logging for each situation?

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,15 +2,21 @@
...
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.debug("login method called with login: {}", login);
         User user = findByLogin(login);
         if (!user.getPassword().equals(password)) {
+            logger.error("Authentication failed for login: {}", login);
             throw new AuthenticationException("Username or password are incorrect");
         }
+        user.logUserLogin();
```
olekskov, line 19:
```suggestion
        logger.info("User logged in - User ID: {}, Login: {}", user.getUserId, user.getLogin);
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #673, opened 2023-06-17
Configured logger, created log file and implemented some logger methdos

vaniii1:


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+app.log
```
Ivan95kos, line 4:
```suggestion
logs/*
```

logs/app.log
```diff
@@ -0,0 +1,6 @@
+2023-06-17 18:07:09,260 INFO mate.academy.service.AuthenticationServiceImpl:13 login method was called. Params: login = bob
```
Ivan95kos, line 1: PR should be without log-file

Ivan95kos (changes requested):

logs/app.log
okuzan: not fixed

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("completeOrder method was called. Params: userId = {}", userId);
```
okuzan, line 15: method is called are more suitable for debug level, change in two places

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,21 @@
...
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
+            OrderService orderService = new OrderServiceImpl();
+            orderService.completeOrder(user.getUserId());
         } catch (AuthenticationException e) {
-            e.printStackTrace();
-            return;
```
okuzan, line 18: Be careful, by deleting return you changed the actual logic of application, which should not be done in this task

okuzan (changes requested):

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
-        // NOTE: In production ready code this order identifier should be generated by DB
-        // For test purpose we simplify this and return dummy data
         order.setOrderId(1L);
         return order;
     }
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
```
overpathz, line 23: Before this we can add more logging like:
logger.info("Fetching products from shopping cart for user: {}", userId);

overpathz (approved):

PR #682, opened 2023-06-30
added solution

Serhii-Nosulia:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Products from DB successfully retrieved. Params: userId = {}", userId);
```
olekskov, line 31: it's better to add info regarding retrieved data details in order to be able to check discrepancy between what user ordered and what obtained in case of bug report

olekskov (approved):
Good job overall, approved

PR #708, opened 2023-07-30
added logger for app

knyazkyylubomir:


olekskov (approved):
Good job, approved.
But further in general i would not recommend extracting logging messages to class constants. As most probably they will be used just once and will draw attention from business-logic class constants/variables

PR #710, opened 2023-08-03
completed the main tasks

ivan0dyatlyukkk:


pom.xml
```diff
@@ -19,6 +19,12 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.20.0</version>
+        </dependency>
+
```
Denis-Balako, line 27: don't forget to remove empty lines

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can't authenticate the user" + e);
```
Denis-Balako, line 21: In the theory we saw, here used the "error level" log. But from my point of view, we should use the "warn level", because that's a common mistake by user and it don't stops or put down a whole logic/server. 
But let's give mentors a decision.

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -25,6 +30,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
         // TODO: add log message about successful fetched data from DB
+        logger.info("The user's orders data was successful fetched from DB");
```
Denis-Balako, line 33: add userId here too, for better debugging in the future.

Denis-Balako (changes requested):
Great work! Fix minor issues, it should be fast!

src/main/java/mate/academy/Main.java
ivan0dyatlyukkk: Ohh, yes, it's my bad, I didn't spot it/ You are right!

Denis-Balako (approved):
Great job!

Vasul-Pankiv (approved):

PR #724, opened 2023-08-08
Initial commit

BlueVioletTeti:


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,14 +2,18 @@
...
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Method login was called. Params: login = {}", login);
         User user = findByLogin(login);
         if (!user.getPassword().equals(password)) {
-            throw new AuthenticationException("Username or password are incorrect");
+            logger.error("Username or password are incorrect");
```
ConfusedSatlan, line 16: You throw exception in main class in try catch block and there you already write error info to log file, so you must not to delete throw new exception and write error again. And in catch block you can write smth like looger.error("....." + e.getMessage());

ConfusedSatlan (changes requested):
Good solution. But there is one improvement.

ConfusedSatlan (approved):

Skarbey (approved):

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Method login was called. Params: login = {}", login);
```
vovasalyha, line 13: Logging that some method was called is more suitable for debug level, as usually you don't want to overwhelm your logs with unnecessary information.

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called. Params: userId = {}", userId);
```
vovasalyha, line 15: Same here.

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name = "mane.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
vovasalyha, line 23: image

vovasalyha (changes requested):

vovasalyha (approved):

PR #732, opened 2023-08-09
logger implemented with configs v 1.0

Skarbey:


vovasalyha (approved):
Approving, but please change how you log errors (you should pass error to log) and change logging of method calls to debug level.

PR #733, opened 2023-08-09
Ihor kalaur

IhorKalaur:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("completeOrder method was called. Params: userId = {}", userId);
```
vovasalyha, line 15: Same here.

vovasalyha (approved):
Approving, but logging information about calling some method is not want you want to do most of the time, for this purpose `debug` level is more suitable.

PR #741, opened 2023-08-16
add logger

silchenko-arsen:


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,16 @@
...
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager
+            .getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Method login was called.");
```
olekskov, line 14: You don't have login logged neither here nor in main method. You won't find required log if user complains on login procedure. Please add where more convenient for you

olekskov (changes requested):

okuzan (approved):

PR #745, opened 2023-08-29
task solution - pom.xml - dependency added

NedashkivskijTV:
- and settings file log4j2.xml is created
- the logger object was added and the logging algorithm was used in classes Main, AuthenticationServiceImpl, OrderServiceImpl
- .gitignore file changed to ignore log file when syncing repository


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+**/logs/app.log
```
Rommelua, line 4:
```suggestion
logs/*
```

Rommelua (approved):

PR #765, opened 2023-09-20
'Logger' task is implemented

Alexsandr-Yablunovskyi:


OlehReex (approved):
Completely agree with your solution!

liliia-ponomarenko (approved):
LGTM 🔥

PR #766, opened 2023-09-20
HW logger

OlehReex:


validatorr (approved):
LGTM! The only thing left is to clear all comments

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,16 @@
...
 import mate.academy.model.User;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
         //TODO: add corresponding log message about method login was called
```
NazarSmith228, line 13: you can clear TODOs now

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,16 @@
...
+        logger.info("login method was called. Params: login = {}", login);
```
NazarSmith228, line 14: it is rather a debug log

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,16 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
         // TODO: add log message about method completeOrder was called
+        logger.info("completeOrder method was called. Params: userId = {}", userId);
```
NazarSmith228, line 16: it is rather a debug log

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+                    <Pattern>%d %p %c:%L %m%n</Pattern>
+            </PatternLayout>
+        </File>
+    </Appenders>
+    <Loggers>
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
```
NazarSmith228, line 18: are you sure that the root level should be `error`?

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -25,6 +30,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
         // TODO: add log message about successful fetched data from DB
+        logger.info("Successfully fetched data from the database. Params: userId = {}", userId);
```
NazarSmith228, line 33: it is rather a debug log

also you need to think of the log content - I would rather log smth like `Successfully fetched data for user: {}`

NazarSmith228 (changes requested):

NazarSmith228 (approved):
LGTM!

PR #768, opened 2023-09-20
Updated logger

AlexRosenkreuz:
Configured logger and completed TODO


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,16 @@
...
 import mate.academy.model.User;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
         //TODO: add corresponding log message about method login was called
```
Parashchuk, line 13: Let's delete the TODO comments in all files where you've done

Parashchuk (approved):

Parashchuk:
Good solution, just minor improvements needed 👍

pom.xml
```diff
@@ -25,6 +25,18 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+            <version>2.11.2</version>
+        </dependency>
```
NazarSmith228, line 33: why do you need this dependency?

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Login() method was called with login={}", login);
```
NazarSmith228, line 13: it is rather a debug log

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("completeOrder() method was called with userId={}", userId);
```
NazarSmith228, line 15: it is rather a debug log

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Data was successfully fetched from DB");
```
NazarSmith228, line 31: it is rather a debug log

also do you need to specify `userId` ?

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+                <Pattern>%d %p %c:%L %m%n</Pattern>
+            </PatternLayout>
+        </File>
+    </Appenders>
+    <Loggers>
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
```
NazarSmith228, line 18: why `error`?

NazarSmith228 (changes requested):

pom.xml
AlexRosenkreuz: It's redundant. I was testing smth and forgot to take that off

src/main/resources/log4j2.xml
AlexRosenkreuz: I used error so that I won't receive all the default logs of program execution and will clearly see my custom logs and how they worked out.

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Processing your login: {}", login);
```
NazarSmith228, line 13: you should use `logger.debug` here, as I mentioned earlier

in addition -  `your login` - bad statement, since our app doesn't know anything about him/her/etc
you should better say `Processing login`

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Processing order for user: {}", userId);
```
NazarSmith228, line 15: you should use `logger.debug` here, as I mentioned earlier

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Successfully received data for user: {}", userId);
```
NazarSmith228, line 31: you should use `logger.debug` here, as I mentioned earlier

src/main/resources/log4j2.xml
NazarSmith228: yeah, I understand this

however the root level is usually `info` in order to understand how the application's lifecycle and not just receiving error logs

NazarSmith228 (changes requested):
One last step :)

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
AlexRosenkreuz: Oooh right. This is what youve meant. Misunderstood you the first time :)

Andriy-dz (approved):

PR #770, opened 2023-09-20
implemented logger

vsych321:


anigilator479 (approved):

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("User couldn't authenticate");
```
anigilator479, line 21: i made this using logger.trace("User couldn't authenticate", e) , but i am not really sure about it

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.debug("completeOrder method was called. Params: userId={}", userId);
```
anigilator479, line 15: if you use logger.debug() this logg won't appear in logs

NazarSmith228: no, actually it is correct to use `debug` level for such logs

the appearence should be configured via configuration file - usually the root level is set to `info`, but when you wan't to debug smth, it is set to `debug`

pom.xml
```diff
@@ -22,10 +22,16 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13.1</version>
```
NazarSmith228, line 25: you don't need this for your PR

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.debug("login method was called. Params: login={}", login);
```
NazarSmith228, line 13: try to pick up the most compact and clear messages for logs

in this case I would log smth like: `Performing login for {}`

src/main/java/mate/academy/service/OrderServiceImpl.java
NazarSmith228, line 15: try to pick up the most compact and clear messages for logs

in this case I would log smth like: `Completing order for {}`

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Data was successfully fetched from DB");
```
NazarSmith228, line 31: it is rather a `debug` log

also, should you specify `userId`?

NazarSmith228 (changes requested):

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Error during using logger", e);
```
NazarSmith228, line 21: weird message :)

`Error during user login`

pom.xml
NazarSmith228: @vsych321 not fixed :)

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.debug("login method was called. Performing login for={}", login);
```
NazarSmith228, line 13: minor

on real environments logs are gathered and stored in some files on some machines

this implies some memory usage, since these files are usually large (20 GB, 30 GB etc.)

this means that you should try to keep your log messages as short as possible but at the same time they should be self-explanatory to bring some value

considering that I would suggest
```suggestion
        logger.debug("Performing login for={}", login);
```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.debug("Completing order method was called with userId={}", userId);
```
NazarSmith228, line 15: as I mentioned above - you should try to keep your logs as short as possible
```suggestion
        logger.debug("Completing order for userId={}", userId);
```

NazarSmith228 (changes requested):
One last step :)

NazarSmith228 (approved):
LGTM

PR #776, opened 2023-09-20
Implemented all the methods and classes

Cayman4010:


KDanil2000 (approved):
LGTM!!

pom.xml
```diff
@@ -22,9 +22,19 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13.1</version>
```
NazarSmith228, line 25: you don't need to update JUnit version in scope of this PR

src/main/resourses/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+                <Pattern>%d %p %c %m%n</Pattern>
+            </PatternLayout>
+        </File>
+    </Appenders>
+    <Loggers>
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
```
NazarSmith228, line 18: why `error`?

in practice most applications have the root level set to `info`

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Method login was called. Login: {}", login);
```
NazarSmith228, line 13: it is better to use `logger.debug()` here

also think of the log message - it should be compact and clear

I would log smth like `Performing login for {}`

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called. UserId: {}", userId);
```
NazarSmith228, line 15: it is better to use `logger.debug()` here

also think of the log message - it should be compact and clear

I would log smth like `Completing order for user: {}`

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Data from database was fetched successful");
```
NazarSmith228, line 31: it is better to use `logger.debug()` here

should you also mention `userId` in log?

NazarSmith228 (changes requested):

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
Cayman4010: OK. Why debug is better here?

Elena-Bruyako (approved):

PR #778, opened 2023-09-21
Implemented logger

KDanil2000:


SalnikovMarkiian (approved):

Elena-Bruyako (changes requested):
Good job! 
Edit .gitignore file so that your file with logs is NOT pushed to Github

logs/app.log
```diff
@@ -0,0 +1,9 @@
+2023-09-21 17:28:43,423 INFO mate.academy.service.AuthenticationServiceImpl:12 login method was called. Params: login = bob.
```
Andriy-dz, line 1: this folder should not be on GitHub. Let's edit the .gitignore file. And let's write more informative messages for the commits

Andriy-dz (changes requested):

logs/app.log
KDanil2000: I can't delete something from gitHub, i just add folder in .gitignore, that all i can do

pom.xml
```diff
@@ -17,8 +17,17 @@
             https://raw.githubusercontent.com/mate-academy/style-guides/master/java/checkstyle.xml
         </maven.checkstyle.plugin.configLocation>
     </properties>
-
     <dependencies>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+            <version>2.20.0</version>
+        </dependency>
```
NazarSmith228, line 25: you don't need this dependency

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+                <Pattern>%d %p %c:%L %m%n</Pattern>
+            </PatternLayout>
+        </File>
+    </Appenders>
+    <Loggers>
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToConsole"/>
+            <AppenderRef ref="LogToFile"/>
+        </Logger>
+        <Root level="error">
```
NazarSmith228, line 18: root level should be set to `info`

logs/app.log
NazarSmith228: @KDanil2000 try to remove that folder locally

then run your code so that it appears again

and then check how your IDE highlights it - should be something like that:
image
as you can see `HELP.md` is a bit more of a grey color, which means it is ignored by git

NazarSmith228 (changes requested):
Good job! Some minor remarks left

src/main/resources/log4j2.xml
Elena-Bruyako: Not fixed. 
Usually in projects the root level of logs - info

Elena-Bruyako (approved):

PR #782, opened 2023-09-24
Added dependency log4j-core, file log4j2.xml, and created Logger in …

Vlad1507:
…classes AuthenticationServiceImpl, OrderServiceImpl, and Main.


Rommelua (changes requested):
image

aaananas (approved):

PR #793, opened 2023-10-12
Hw logger

kaptetos:


.gitignore
```diff
@@ -1,3 +1,4 @@
+my_log_file.log
```
kozhukhovsky, line 1: What if you add a date to the naming pattern and it will be a different file?
It's best to add the log folder here so that any files there don't end up in the repository.

kozhukhovsky (changes requested):
Add logs to the service files

kozhukhovsky (changes requested):
I don't see the folder in gitignore.
And I also don’t see any corrections to my comment regarding services.

okuzan (changes requested):
you forgot about the OrderServiceImpl

olekskov (approved):

PR #801, opened 2023-10-20
Solution for the Logger task

Stelmaschyk:


logs/app.log
```diff
@@ -0,0 +1,2 @@
+2023-10-20 11:24:09,115 INFO mate.academy.service.OrderServiceImpl:15 CompleteOrder method was called
+2023-10-20 11:24:09,119 INFO mate.academy.service.OrderServiceImpl:27 GetAllProductsFromShoppingCartData method was called. Data from DB were successful fetched
```
okuzan, line 2: logs shouldn't be committed - logs/* folder should be added to gitignore

okuzan (changes requested):

olekskov (approved):

PR #808, opened 2023-10-29
Edited .gitignore, added logs in classes, created package-info.java i…

Monsterhunter525:
…n all packages, refactored pom.xml file(with help of chatGPT).


src/main/java/mate/academy/Main.java
```diff
@@ -7,14 +10,33 @@
...
+    public static void main(final String[] args) {
+        AuthenticationService authenticationService =
+                new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            LOGGER.severe("An authentication exception occurred: "
+                    + e.getMessage());
```
Sanderfuck, line 39: You don`t need concatenate string and exception, this method can receive two args error(String, e)
And use method error if you log in catch block

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,21 +2,49 @@
...
-    public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+    public User login(final String login, final String password)
+            throws AuthenticationException {
+        LOGGER.info("Method login was called with login: " + login);
         User user = findByLogin(login);
         if (!user.getPassword().equals(password)) {
-            throw new AuthenticationException("Username or password are incorrect");
+            throw new AuthenticationException(
+                    "Username or password are incorrect");
```
Sanderfuck, line 33: What about log.error?

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,21 +2,49 @@
...
+     * @param login the login of the user
+     * @return the user with the specified login
+     */
+    private User findByLogin(final String login) {
         User user = new User(login, "1234");
-        // this user identifier should be set by DB. We will use dummy data for this example
+        // this user identifier should be set by DB.
+        // We will use dummy data for this example
         user.setUserId(2L);
         return user;
```
Sanderfuck, line 49: You can log.info for receiving parameters in method

Sanderfuck (changes requested):

src/main/java/mate/academy/Main.java
```diff
@@ -1,20 +1,40 @@
...
-    public static void main(String[] args) {
-        AuthenticationService authenticationService = new AuthenticationServiceImpl();
+import java.util.logging.Logger;
+
+public final class Main {
+    /**
+     * The logger for the Main class.
+     */
+    private static Logger logger = Logger.getLogger(
+            Main.class.getName());
```
sokimaaa, line 19: make logger the private static final. 
getLogger() can accept the Main.class directly
```suggestion
    private static final Logger logger = Logger.getLogger(
            Main.class);
```

src/main/java/mate/academy/Main.java
```diff
@@ -1,20 +1,40 @@
...
 import mate.academy.model.User;
 import mate.academy.service.AuthenticationService;
 import mate.academy.service.AuthenticationServiceImpl;
 import mate.academy.service.OrderService;
 import mate.academy.service.OrderServiceImpl;
 
-public class Main {
-    public static void main(String[] args) {
-        AuthenticationService authenticationService = new AuthenticationServiceImpl();
+import java.util.logging.Logger;
```
sokimaaa, line 12: using java logger is not very good solution. 
try to use `logback` or `log4j2` instead
```suggestion
import java.util.logging.Logger;
```

sokimaaa (changes requested):

src/main/java/mate/academy/Main.java
Monsterhunter525: [ERROR] C:\Users\User\IdeaProjects\jv-logger\src\main\java\mate\academy\Main.java:18:33: Name 'logger' must match pattern '^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$'. [ConstantName]
Audit done.
when i do it final it asks me to do constant, and when i do constant other mentor is saying to do it in lowercase, and what shal i  do?))) and when i do it like on a example i must wrap it in toString, and i still have the same error

Sanderfuck (approved):

PR #810, opened 2023-10-31
Added logger

HovorukhaBohdan:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called");
```
kshuryhin, line 15: That would be better to provide also arguments, if no sensitive data, to the log message. Like, "method x was called with argument y"

kshuryhin (changes requested):
LGTM, just one minor point to consider

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -12,7 +12,7 @@ public class OrderServiceImpl implements OrderService {
 
     @Override
     public Order completeOrder(Long userId) {
-        logger.info("Method completeOrder was called");
+        logger.info("Method completeOrder was called with userId: " + userId);
```
kshuryhin, line 15: I guess we can use `{}` for parameters injecting, but that's minor

kshuryhin (approved):
LGTM

PR #815, opened 2023-11-06
Finished

Oleksa-32:


okuzan (changes requested):
Debug log should be info and vice versa
Something is wrong with your project structure
image

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called. Params: userId={}", userId);
```
okuzan, line 15: 'method was called' type of logs are usually debug level

okuzan (approved):
Good job

PR #825, opened 2023-11-08
create log messages

denysokhrymovych:


vadymhrnk (approved):
Everything looks good to me, but I think it would be better to make messages in `logger.info()` and `logger.error()` as constants.

kshuryhin (changes requested):
Good job, however locally I do not see any logs in the console output or file, could you provide any screenshots?

bhdnchui (approved):
Good job!

PR #835, opened 2023-11-12
RealizationOfLogging

forzen123212:


src/main/java/mate/academy/Main.java
```diff
@@ -6,16 +6,19 @@
 import mate.academy.service.AuthenticationServiceImpl;
```
BolsunovaNataliia, line 6: Good job!

BolsunovaNataliia (approved):

pom.xml
```diff
@@ -21,9 +21,15 @@
...
-            <version>4.12</version>
+            <version>4.13.1</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.21.1</version>
+        </dependency>
+
```
vovasalyha, line 32:
```suggestion
```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Method getAllProductsFromShoppingCart was called. Params = {}", userId);
```
vovasalyha, line 27: This method will be called very frequently in real app, so it's better to use DEBUG level.

src/main/java/mate/academy/Main.java
```diff
@@ -6,16 +6,19 @@
...
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
-        User user;
+        User user = new User("bob","1234");
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
-            return;
+            logger.debug("Method login was called. Params login = {}", "bob");
```
vovasalyha, line 21: I think some other message should be here, as an exception was thrown. Also, think about other log level.

vovasalyha (changes requested):
Where is your xml configuration of appenders? Also see a couple of comments

bhdnchui (approved):
Good job!

PR #876, opened 2024-01-19
Solution

suprun-anna:
added dependency for log4j to pom.xml, added configuration file, added logging to methods


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
 import java.util.List;
 import mate.academy.model.Order;
 import mate.academy.model.Product;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
```
sokimaaa, line 11: Wrong class -> logger will log wrong data
```suggestion
    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
```

sokimaaa (changes requested):

sokimaaa (approved):

PR #878, opened 2024-01-22
hm for review

vdbond:


kozhukhovsky (changes requested):
Where is logger config file?

kozhukhovsky (approved):

PR #900, opened 2024-03-11
fixed

LizaBalshv:


logs/app.log
olekskov, line 1: log file shouldn't be commited, you added .logs/* to .gitignore not to have this file

olekskov (approved):
Good job

PR #914, opened 2024-03-25
Added implementation for task Logger...

nklimovych:
Added dependency for log4j to pom.xml file. Also created config file log4j2.xml, and added configuration. All todo were replaced by logger. Cleared all comments.


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+logs/*
```
vikkushnir, line 4: Probably, you need here additional line
```suggestion
logs/*

```

pom.xml
```diff
@@ -24,6 +24,12 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.23.1</version>
+        </dependency>
+
```
vikkushnir, line 32:
```suggestion
```

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="INFO">
+    <Appenders>
+        <Console name="Console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+        </Console>
+        <File name="File" fileName="logs/app.log">
+            <PatternLayout>
+                <Pattern>%d %p %c: %line %msg%n</Pattern>
```
vikkushnir, line 9: you can use here %L instead of %line, just in case, not necessary
```suggestion
                <Pattern>%d %p %c:%L %msg%n</Pattern>
```

vikkushnir (approved):
Good job!) just some minor things

andrii-hoienko (changes requested):
Please resolve all student comments

src/main/resources/log4j2.xml
nklimovych: Thanks for your suggestion but I made it on purpose) 
In my opinion, this option is significantly more informative 😊

pom.xml
nklimovych: I don't understand how I missed that 😅

src/main/resources/log4j2.xml
karma-o: I'd also add a level and date and time formatted

karma-o (approved):

PR #917, opened 2024-03-26
Solution was added

slizko1:


azzilian (changes requested):
All good - just Edit the .gitignore file so that your file with logs is NOT pushed to GitHub - add *.log like this

.idea/*
*.iml
*.log
target/*

PR #918, opened 2024-03-26
Add logger to project

TonyH277:


src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,20 @@
...
+    <Loggers>
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="Console"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="Console"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
ValentynSharshon, line 20:
```suggestion
</Configuration>

```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("getAllProductsFromShoppingCart success for: userId - {}", userId);
```
ValentynSharshon, line 27: I think the INFO logging level should be used here.

ValentynSharshon (approved):
LGFM! Just minor comments.

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,21 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
+        OrderService orderService = new OrderServiceImpl();
         User user;
         try {
-            user = authenticationService.login("bob", "1234");
+            user = authenticationService.login("bob", "1254");
+            orderService.completeOrder(user.getUserId());
```
karma-o, line 21: about this `completeOrder()`, would be better to leave it outside the try block, I'd always strive to keep the try blocks shorter, to make it more obvious where the exception might be thrown, and that's and overall good practice

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,21 @@
...
         } catch (AuthenticationException e) {
-            e.printStackTrace();
-            return;
+            logger.error("Username or password are incorrect" + e);
```
karma-o, line 23: to make it more clear
```suggestion
            logger.error("Login failed. Username or password are incorrect" + e);
```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.debug("CompleteOrder method was called with params: userId - {}", userId);
```
karma-o, line 15: I think this one could be info because `complete order` seems important enough

src/main/java/mate/academy/service/OrderServiceImpl.java
karma-o: I mean it could, but I think It is fine both ways, so up to you

src/main/resources/log4j2.xml
karma-o: looks like it already has a new line at the end of it, it usually shows a red sign if it does not

karma-o (approved):
Left some minor comments and approved, please take a look

PR #919, opened 2024-03-26
logger configured and added

azzilian:
All 3 TODO completed
e.printStackTrace() - replaced with logger.error


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,36 @@
...
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
+        logger.info("Method getAllProductsFromShoppingCart was called");
         // NOTE: In production ready code this method should fetch data from DB
         // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Returning list of All products for userId : {}", userId);
```
VlasenkoYevhenii, line 36: The TODO: was about to log the fact the method has been called )
But I don't know if it matters so it's up to you =)

VlasenkoYevhenii (changes requested):
GJ, buddy!
Approved but with a minor comment

src/main/java/mate/academy/service/OrderServiceImpl.java
azzilian: I got  logger.info("Method getAllProductsFromShoppingCart was called"); at the line above)

VlasenkoYevhenii: Also about your config xml file: as I see, we don't have any System.out,println() to replace with the corresponding logger message, so do we really need configuration for the Console out. IDK, just a suggestion =)

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,20 @@
...
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
+            logger.info("Password check succesfull returning user...");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can`t login with user: {}", e.getMessage());
```
karma-o, line 22: just pass the exception, not the message
```suggestion
            logger.error("Can`t login with user: {}", e);
```

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,20 @@
...
 
 public class Main {
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
+            logger.info("Password check succesfull returning user...");
```
karma-o, line 20: this is actually a thing we want to avoid because this is kind of confidential data, better to avoid anything that has a relation to the password, even a simple confirmation like that can be beneficial for hackers

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,12 +2,17 @@
...
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Method login was called");
         User user = findByLogin(login);
+        logger.info("Attempt to Log In using login: {}", login);
```
karma-o, line 15: the first message could be dropped in that case, because second one does essentially the same thing, also this looks more like a `trace` or `debug` level to me

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,36 @@
...
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called");
+        if (userId == null) {
+            logger.error("Incoming data for userId is null");
```
karma-o, line 17: you log error but essentially ignoring the fact that it is null, I wouldn't complicate things with this null check

src/main/java/mate/academy/service/OrderServiceImpl.java
karma-o: agree, the message should be about succesfull fetch from the DB, and not about the method call or that it returns some value, so let's do that

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToConsole"/>
+            <AppenderRef ref="LogToFile"/>
+        </Logger>
+        <Root level="error" additivity="false">
+            <AppenderRef ref="LogToConsole"/>
+            <AppenderRef ref="LogToFile"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
karma-o, line 23:
```suggestion
</Configuration>

```

karma-o (changes requested):
some minor comments

src/main/java/mate/academy/service/OrderServiceImpl.java
azzilian: > Also about your config xml file: as I see, we don't have any System.out,println() to replace with the corresponding logger message, so do we really need configuration for the Console out. IDK, just a suggestion =)

But its in the task itself Configure the logger in this project. Add appenders for File and Console

src/main/java/mate/academy/Main.java
karma-o: Actually we do not need those braces too, my bad, did not see that at the first look
we need them when we want to add something to the message like, but the exceptions can be passed without it

karma-o (approved):

PR #922, opened 2024-03-26
implement logger

HAIOVYI:


fedaktaras (approved):
Fine.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+log4j2.xml
```
karma-o, line 4: you should put the `logs`  folder to `.gitignore` to not push the `.log` files of your application to github, `log4j2.xml` on the other hand should actually be pushed because it is a config file which should be in our repository, and it is still here, because you haven't specified the directory where it is or *.
Next time you can verify yourself what files were pushed to git and what did not, by clicking `Files` in your PR

karma-o (changes requested):
check gitignore

.gitignore
HAIOVYI: omg. thx

PR #927, opened 2024-03-26
build success

MolchanovAlexander:


AnyaRatusnyak (approved):
Well done!!

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+logs/*
```
karma-o, line 4: Always nice to have an empty line at the end of the file
```suggestion
logs/*

```

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
 import mate.academy.service.AuthenticationServiceImpl;
 import mate.academy.service.OrderService;
 import mate.academy.service.OrderServiceImpl;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class Main {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
```
karma-o, line 13: What if we have another exception with some other service involved? We'll have a misguiding log that `AuthernticationServiceimpl` threw some exception when in fact it has nothing to do with it. We pass a class where this logger is located to the parameter, not what the exception is, or about what is this log
```suggestion
    private static final Logger logger = LogManager.getLogger(Main.class);
```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
 import java.util.List;
 import mate.academy.model.Order;
 import mate.academy.model.Product;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
```
karma-o, line 11: same here
```suggestion
    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
```

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="Console"/>
+            <AppenderRef ref="LogToFile"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="Console"/>
+            <AppenderRef ref="LogToFile"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
karma-o, line 23:
```suggestion
</Configuration>

```

karma-o (changes requested):

src/main/java/mate/academy/service/OrderServiceImpl.java
MolchanovAlexander: My bad.

karma-o (approved):

PR #931, opened 2024-03-26
added logger and necessary changes

AnyaRatusnyak:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Can't login", e);
```
VolandevlodD, line 21: You should make error messages more informative, and you could also consider passing some parameters to the messages.
```suggestion
          logger.error("Failed to login user with username '{}' "
                    + "due to authentication error: {}", "bob", e.getMessage());
```

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Login method was called. Params: login = {}", login);
```
VolandevlodD, line 13: In my humble opinion, logs that are not important should be logged at the debug level. However, the decision is still yours.
```suggestion
        logger.debug("Login method was called. Params: login = {}", login);
```

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called. Params: userId = {}", userId);
```
VolandevlodD, line 15: here also

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Successfully fetched data from DB. Params: userId = {}", userId);
```
VolandevlodD, line 27: And here, I believe we should use the debug level of logging.

VolandevlodD (approved):
LGFM minor comments

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
andrii-hoienko: Agree

src/main/java/mate/academy/service/OrderServiceImpl.java
andrii-hoienko: Agree

src/main/java/mate/academy/service/OrderServiceImpl.java
andrii-hoienko: Disagree

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" additivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
andrii-hoienko, line 23:
```suggestion
</Configuration>

```

src/main/java/mate/academy/Main.java
andrii-hoienko: Up to you, but it would be better to make the message more informative, like suggested for example

andrii-hoienko (changes requested):
Good job, but some improvements required

AnyaRatusnyak:
Thank you)

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,20 @@
...
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Failed to login user with username '{}' "
+                    + "due to authentication error: {}", "bob", e.getMessage());
```
karma-o, line 22: we should also pass the exception to the logger to maintain it there
```suggestion
            logger.error("Failed to login user with username '{}' "
                    + "due to authentication error: {}", "bob", e.getMessage(), e);
```

karma-o (approved):
Approved with a minor comment, please take a look :)

PR #933, opened 2024-03-27
implemented logger in code

kaiiiseeel:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Data successfully fetched from DB");
```
oksana-miazina, line 31: We can also add `userId` to the message

oksana-miazina (approved):
Clean solution!

src/main/java/mate/academy/service/OrderServiceImpl.java
andrii-hoienko: agree

andrii-hoienko (changes requested):
Very nice, left minor comment

karma-o (approved):

PR #944, opened 2024-04-12
added logger

nastasiiaKulinich:


src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+        <Logger name="mate.academy" level="info" addictivity="false">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Logger>
+        <Root level="error">
+            <AppenderRef ref="LogToFile"/>
+            <AppenderRef ref="LogToConsole"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
okuzan, line 23:
```suggestion
</Configuration>

```

okuzan (changes requested):
in Main there is a printStackTrace();, replace with log.error

olekskov (approved):

PR #953, opened 2024-04-26
jv-logger task complete

YanaBN:
log4j2.xml was added, application work was checked, checklist followed


sarakhmen (changes requested):
Let's improve your solution :)

Elena-Bruyako (approved):

PR #958, opened 2024-05-02
The task has done according 'README' file.

NekitCool1:


Elena-Bruyako (changes requested):
Provide log4j2.xml file in PR

.gitignore
```diff
@@ -1,3 +1,3 @@
 .idea/*
 *.iml
-target/*
+target/*
```
sarakhmen, line 3: please also add logs/ here so that logs aren't pushed to github and added to git history

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Login method was called. Params: login - {}", login);
```
sarakhmen, line 13: it's better to use debug level instead of info when you want to trace method calls

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("Method completeOrder was called. User ID = {}", userId);
```
sarakhmen, line 15: same issue here

sarakhmen (changes requested):
Let's improve your solution :)

MentorMA (approved):

PR #962, opened 2024-05-16
added loggers

Nazar090:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,17 @@
...
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger =
+            LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
         // TODO: add log message about method completeOrder was called
+        logger.info("completeOrder method was called.");
```
okuzan, line 17: 'method was called' type of logs is more suitable for debug level

okuzan (approved):

PR #968, opened 2024-05-24
Configured the logger:

ipantazi:
Added dependency to the `pom.xml` file.
Created a file `log4j2.xml`.
Added "*.log" exception to the `.gitignore` file.
Replaced "e.printStackTrace()" with a logger method. All TODO instructions have been completed..

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationServiceImpl.java
+++ b/src/main/java/mate/academy/service/AuthenticationServiceImpl.java
@@ -10,7 +10,7 @@ public class AuthenticationServiceImpl implements AuthenticationService {
 
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        logger.info("Method login was called. Params: login={}", login);
+        logger.debug("Method login was called. Params: login={}", login);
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Method login was called. Params: login={}", login);
```
sarakhmen, line 13: usually it's better to use debug level in order to log method calls

sarakhmen (changes requested):

Elena-Bruyako (approved):

PR #972, opened 2024-06-01
Jv logger

Vadym-Ozi:
Created logger. Added logger to methods


src/main/java/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
...
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+        </Console>
+        <File name="LogToFile" fileName="logs/app.log">
+            <PatternLayout>
+                <Patern>%d %p %c %m%n</Patern>
+            </PatternLayout>
+        </File>
+    </Appenders>
+    <Loggers>
+        <Logger name="mate.academy" level="info" additivity="false">
```
Elena-Bruyako, line 14:
```suggestion
        <Logger name="mate.academy" level="debug" additivity="false">
```

src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="info">
```
Elena-Bruyako, line 2: It looks like this file should be exists because you have already one

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #978, opened 2024-06-19
Configured the logger and did TODO tasks

YuliiaNisha:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,27 +4,27 @@
...
 
     private List<Product> getAllProductsFromShoppingCart(Long userId) {
-        // NOTE: In production ready code this method should fetch data from DB
-        // For test purpose we simplify this method and return dummy data
         Product iphone = new Product("iPhone X", BigDecimal.valueOf(1199));
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("Got the products from the shopping cart. User: " + userId);
```
sarakhmen, line 27: let's use info level here. We usually use info log level to keep track of retrieved data from DB. Debug level is usually used for testing purposes, tracking method calls etc

sarakhmen (changes requested):

pom.xml
```diff
@@ -24,6 +24,16 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
```
Elena-Bruyako, line 28: remove this dependency

Elena-Bruyako (approved):
Good job! Just don't add dependency, that you actually don't need

PR #988, opened 2024-07-08
jv-logger task solution

SvNadorozhnyi:


src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.debug("The data successfully fetched from DB by userId - {}", userId);
```
sarakhmen, line 31: for such an information we usually use info() method instead of debug()

sarakhmen (changes requested):

Elena-Bruyako (approved):

PR #995, opened 2024-08-03
Hw logger

fetis-off:


logs/app.log
```diff
@@ -0,0 +1,12 @@
+2024-08-03 11:37:47,724 INFO mate.academy.service.AuthenticationServiceImpl Login method was called
```
okuzan, line 1: logs folder and its content should be in the PR

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("comleteOrder method was called. Params: userId = {}", userId);
```
okuzan, line 15: method was called type of logs is debug level, not info

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Login method was called. Params: login = {}", login);
```
Elena-Bruyako, line 13:
```suggestion
        logger.debug("Login method was called. Params: login = {}", login);
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1005, opened 2024-08-25
installed log4j2 and added logging to the app

NikitaLafinskiy:


src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,19 @@
...
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Credentials invalid, access unauthorized: ", e);
```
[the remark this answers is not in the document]
Elena-Bruyako: fix comment

log4j2.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Elena-Bruyako, line 1: this file should be placed in resources folder

src/main/java/org/example/Main.java
```diff
@@ -0,0 +1,7 @@
+package org.example;
+
+public class Main {
```
Elena-Bruyako, line 3: Why do you need this class?

Elena-Bruyako (changes requested):

src/main/java/org/example/Main.java
NikitaLafinskiy: I am not sure where it came from to be fair 
Though it was included in the source directory

NikitaLafinskiy:
My "main" method's try/catch block currently looks like this, perhapse it did not get added in the previous commit:
try {
            logger.error("Credentials invalid, access unauthorized: ", e);
            return;
        }

Elena-Bruyako (approved):

PR #1015, opened 2024-09-09
commit logger

nplastun:


Elena-Bruyako (changes requested):
Didn't find the configuration according to the task description

Configure the logger in this project. Add appenders for File and Console

Elena-Bruyako (approved):

PR #1018, opened 2024-09-16
Logger

dlvsn:


src/main/java/mate/academy/Main.java
```diff
@@ -6,16 +6,19 @@
...
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
-        User user;
+        User user = null;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
-            return;
```
okuzan, line 18: deleting return changes logic of the problem

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,10 +2,15 @@
...
 import mate.academy.model.User;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
+        logger.info("Method login was called with login-> '{}'", login);
```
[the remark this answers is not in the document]
okuzan: method was called more suitable for DEBUG level

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+*.log
```
okuzan, line 4:
```suggestion
*.log
logs/*
```

okuzan (changes requested):

PR #1035, opened 2024-10-27
commit task

Oleksii21th:


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,13 +2,18 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Login method was called! Params login: {}", login);
```
slade13, line 13: Avoid logging sensitive information such as login credentials. This could be a security risk.

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,20 @@
...
 
 public class Main {
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
+            logger.info("User {} logged in successfully.", user.getLogin());
```
slade13, line 20: Avoid potential NullPointerException. Add null check before calling 'getLogin()' on 'user' object.

src/main/java/mate/academy/Main.java
```diff
@@ -6,15 +6,20 @@
...
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Login failed for user bob with message: {}", e.getMessage());
```
slade13, line 22: Avoid hardcoding user credentials ('bob' in this case). Consider using an argument or variable instead.

slade13, line 22: Use logger.error() method to log both the custom message and the stack trace of the exception for better debugging.

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -24,7 +28,7 @@ private List<Product> getAllProductsFromShoppingCart(Long userId) {
         Product macBook = new Product("MacBook Air 2020", BigDecimal.valueOf(1399));
         Product xiaomi = new Product("Xiaomi 12", BigDecimal.valueOf(499));
         List<Product> products = List.of(iphone, macBook, xiaomi);
-        // TODO: add log message about successful fetched data from DB
+        logger.info("Product data returned successfully for testing.");
```
slade13, line 31: The log message is misleading. It implies that data is fetched from a database, but the data is hard-coded. Consider changing the message to reflect the reality of the operation, and change logger level to `DEBUG`.

src/main/java/mate/academy/service/OrderServiceImpl.java
```diff
@@ -4,11 +4,15 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class OrderServiceImpl implements OrderService {
+    private static final Logger logger = LogManager.getLogger(OrderServiceImpl.class);
+
     @Override
     public Order completeOrder(Long userId) {
-        // TODO: add log message about method completeOrder was called
+        logger.info("CompleteOrder method was called!");
```
slade13, line 15: The log message doesn't provide any context about the user or the operation. Consider including the userId in the log message to improve traceability, and change logger level to `DEBUG`.

src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,13 +2,18 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Login method was called!");
```
slade13, line 13: The logger.info statement should be replaced with logger.debug to avoid cluttering the log files with routine operations.

slade13, line 13: The log message 'Login method was called!' could be more informative. Consider including the user's login in the message to provide more context.

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,29 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class Main {
+    private static final Logger logger = LogManager.getLogger(Main.class);
+
     public static void main(String[] args) {
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
-        User user;
+        User user = null;
```
slade13, line 17: It is unnecessary to initialize the 'user' variable to null as it gets a value assigned in the try block.

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,29 @@
...
         try {
             user = authenticationService.login("bob", "1234");
+            logger.info("User {} logged in successfully.", getUserLogin(user));
```
slade13, line 20: The logger.info() statement uses a method getUserLogin(user) that could return an empty string, this could lead to confusion when analyzing logs as it would not be clear which user the log refers to.

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,29 @@
...
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Login failed for user {} with message: {}",
```
slade13, line 22: The logger.error() statement also uses the method getUserLogin(user) that could return an empty string. It would be more useful to include more specific user information in the error message, if available, to aid in debugging.

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,29 @@
...
+                    getUserLogin(user), e.getMessage(), e);
             return;
         }
         OrderService orderService = new OrderServiceImpl();
         orderService.completeOrder(user.getUserId());
     }
+
+    private static String getUserLogin(User user) {
+        // there is no input username, so I return an empty one
```
slade13, line 31: The comment '// there is no input username, so I return an empty one' is misleading. The function actually returns the username if the user object is not null, and an empty string if the user object is null.

slade13 (approved):
Good job, I left some comments which can be changed but the task is okay.

PR #1052, opened 2024-11-17
logger configured

Nazar1106:


src/main/java/mate/academy/service/AuthenticationServiceImpl.java
```diff
@@ -2,11 +2,16 @@
...
+import org.apache.logging.log4j.Logger;
 
 public class AuthenticationServiceImpl implements AuthenticationService {
+    private static final Logger logger
+            = LogManager.getLogger(AuthenticationServiceImpl.class);
+
     @Override
     public User login(String login, String password) throws AuthenticationException {
-        //TODO: add corresponding log message about method login was called
+        logger.info("Method login was called for user with login: {}", login);
```
Elena-Bruyako, line 14: it's better to use debug level when methods are called

Elena-Bruyako (changes requested):

PR #1061, opened 2024-12-04
final

Nikolya777777777777:


okuzan (changes requested):
?? You deployed bookshop project in jv-logger PR?

Nikolya777777777777:
I answered to your question in jv-chat, please check, I didn't import anything, all these files were there at once, I changed pom.xml, added the log4j2.xml file and created loggers in the classes, added logger.debug, error and info. If I understood your question right)

Elena-Bruyako (changes requested):
Task implemented correctly, but why there are so many files? revert changes in gitignore file

PR #1143, opened 2025-06-28
 implement logging using log4j2:

Mwxyz2022:
- added log4j dependencies to the project
- created the log4j2.xml configuration file
- configured a console appender for real-time logging
- configured a file appender to save logs into logs/app.log
- added the logs/ directory to .gitignore
- implemented logging within AuthenticationServiceImpl for the user login flow
- implemented logging within OrderServiceImpl for the order completion process
- added logging in the Main class to correctly handle authentication exceptions
- replaced e.printStackTrace() with a proper logger.error() call


Mwxyz2022:
@mateacademy-ai-mentor 
I have double-checked, and the log4j2.xml file is present in this pull request. It is located at the correct path: src/main/resources/log4j2.xml.

Could you take another look?

liliia-ponomarenko (approved):
Good job!

PR #1170, opened 2025-10-14
Task for check

JacekKubiak80:


logs/app.log
```diff
@@ -0,0 +1,5 @@
+2025-10-14 13:04:12 [main] INFO  mate.academy.Main - Application started
```
mateuszwojtkowiak, line 1: Your logs should not be pushed to GitHub. It is one of the requirements of this task: "Edit the .gitignore file so that your file with logs is NOT pushed to GitHub". AI bot noticed that earlier and gave you a solution on how to do it: "Fix: add logs/ and *.log to the root .gitignore, then run git rm --cached -r logs/ and commit the removal."

src/main/java/mate/academy/Main.java
```diff
@@ -6,18 +6,27 @@
...
     public static void main(String[] args) {
+        logger.info("Application started");
+
         AuthenticationService authenticationService = new AuthenticationServiceImpl();
         User user;
         try {
             user = authenticationService.login("bob", "1234");
         } catch (AuthenticationException e) {
-            e.printStackTrace();
+            logger.error("Authentication failed: {}", e.getMessage(), e);
```
Wojtas1996PL: You could add here more specific information, for example "Can't login user with login: ".

mateuszwojtkowiak (changes requested):
Great job! Please ensure that your log files are not pushed to GitHub, as required by the task.

MateuszRostek (approved):

